PR #4, opened 2024-10-23
Replacing a vector of paths with PathMap in zipper excl. tracking

marcin-rzeznicki:
Plus a couple of related changes:
- replacing panic with error,
- more precise types.

### `ZipperTracker` is now parametrized by `TrackingMode`

`TrackingMode` serves as a type-level indicator whether we're tracking writes or reads. In Rust you cannot do much with this - but at the very least we don't need to panic at runtime e.g. in `Clone`. Also various constructors throughout the code are aware of what mode they expect, which may be helpful in further development as there is no way to mix read-tracking zippers with write-tracking zippers.

### `BytesTrieMap` is used instad of `Vec` to track locked paths (`SharedTrackerPaths`)

This helps the complexity as checking no longer needs traversing all of the `n` paths and checking their prefixes. 
The replacement uses `BytesTrieMap` as a prefix tree annotated with `IsTracking` values that designate the subtree as locked for writing, or reading (unsigned non-zero counter)

### Replacing panic with error

The `Conflict` type is implemented, serving as a structured error info, and methods that previously panicked (and were poisoning the locks) now return `Result<(), Conflict>` allowing the caller to graciously handle read/write conflicts

## What does not work?

Tests - I have not finished adapting them - they don't eve compile currently. Will fix in a subsequent PR

## Things to scrutinize

Usage of zippers may be imprecise. Please check methods that rely on certain zipper/trie behavior e.g. `Conflict::check_subtree_for_conflict`, `SharedTrackPaths::try_add_writer`, `SharedTrackPaths::try_add_reader`

## I am sorry for

Reformatting a lot of code which makes the PR much bigger. This is my lack of tooling knowledge showing up (I don't know how to turn it off :smile: ). Perhaps we could have some kind of project-wide formatting config to avoid this stylistic inconsistency?

### DIFF
--- a/src/line_list_node.rs
+++ b/src/line_list_node.rs
@@ -1834,7 +1834,7 @@ impl<V: Clone + Send + Sync> TrieNode<V> for LineListNode<V> {
         }
         if key1.len() > key.len() && key1.starts_with(key) {
             let mut new_node = Self::new();
-            unsafe{ new_node.set_payload_0(&key0[key.len()..], self.is_child_ptr::<0>(), ValOrChildUnion{ _unused: () }) }
+            unsafe{ new_node.set_payload_0(&key1[key.len()..], self.is_child_ptr::<0>(), ValOrChildUnion{ _unused: () }) }
             new_node.val_or_child0 = self.take_payload::<1>().unwrap().into();
             debug_assert!(validate_node(&new_node));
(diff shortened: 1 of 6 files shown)


luketpeterson (changes requested):
Looks fine except for the build-breaking change in release mode.  Which should simply be reverted.

PR #16, opened 2025-08-06
Implemented a generic virtual product zipper

imlvts:
This allows using ACT in MORK for product zippers.
Additionally, fixed a few bugs in ACT, and changed the API of ACT to get values easier


Adam-Vandervorst:
We discussed some changes to this on the Zippy a Blaze, this one isn't "generic enough" to support query multi. I'd refrain from renaming until that solution lands, which likely has codepaths parallel to the current generic PZ.

PR #17, opened 2025-09-03
Coro paths

Adam-Vandervorst:
The serialization logic stayed the same, but is now transformed into a coroutine.
Requires nightly and needs to be gated by the `nightly` features flag

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,4 +1,19 @@
 
 [build]
-#rustflags = "-C target-cpu=native -C lto=fat -C embed-bitcode=yes -Z dylib-lto"
+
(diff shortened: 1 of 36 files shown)


luketpeterson:
Ok.  I'm happy to merge this.

1. I created a MORK branch called [use_new_serialize_api](https://github.com/trueagi-io/MORK/tree/use_new_serialize_api) to reflect the API changes I made
2. QUESTION: in the documentation I write, I talk about "`.paths` data" as a shorthand for the target format.  But therefore, consistency would dictate the following renames:
 mod `path_serialization` -> `paths_serialization`
 `path_serialize_sink` -> `paths_serialization_sink`

PR #19, opened 2025-09-17
Dangling paths

luketpeterson:
Change in behavior to ensure paths are always stable (won't be removed as a side effect of another operation)

* Adding checks inside ListNode so Null node ptrs aren't considered an illegal node anymore, and in ByteNode so Cf(None, None) isn't considered illegal
* adding explicit WriteZipper methods to prune and create paths
* adding flag arguments to a handful of existing WriteZipper and PathMap methods to prune explicitly.


luketpeterson:
Main-lines dangling path API and behavior changes

PR #24, opened 2025-09-25
Add WeightedTriemap for path frequency tracking

af271:
Adds WeightedTriemap<V> to PathMap for efficient node-level weight and frequency tracking in trie structures.

WeightedTriemap<V> - Generic wrapper around [PathMap<NodeWeight>]
NodeWeight - Structure with local_count: u64 for frequency tracking

let mut map = WeightedTriemap::<()>::new();
map.increment_count(b"/path");  // Increment count
let weight = map.get_weight(b"/path");  // Get current weight

### DIFF
--- a/src/weighted_trie.rs
+++ b/src/weighted_trie.rs
@@ -19,10 +19,6 @@ impl<V: Clone + Send + Sync + Unpin> WeightedTriemap<V> {
         }
     }
 
-    pub fn new_with_k(_k: usize) -> Self {
-        Self::new()
-    }
-
     pub fn increment_count(&mut self, path: &[u8]) {
         let current_weight = self.inner.get_val_at(path)
             .map(|w| w.clone())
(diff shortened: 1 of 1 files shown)


luketpeterson:
Thanks for submitting this PR.

I'm trying to understand a little bit more about the use case for the `WeightedTriemap`.  What is an algorithm where you'd use this?

Also, I'm wondering if this belongs in the core of the `pathmap` crate, vs. in client code or maybe in the `utils` module.

luketpeterson:
I see.  It seems like you want summary statistics in the map.  We have thought about this kind of feature for a long time, but have come to the conclusion that it is fundamentally incompatible with pathmap's algebraic operators to implement anything in the core.

Within MORK, you have two options.  MORK currently uses a unit type `()` as the PathMap value, so you could replace that with a counter - but I think you'd have a lot less code change and a more maintainable system if you just created a parallel map (as you have done) within your client code to track the frequency metadata.

Based on my current understanding, I don't think this object belongs in the PathMap crate, although it likely has a place at a higher level.

Did I fail to consider anything?  I.e. is there a need for the frequency tracker to interact with pathmap at a low level or use private APIs?

af271:
Where would you suggest putting it? With MORK?
If you put it on the server side, then Python would have to store all the expressions and their respective stats, which would make look ups slow, like giving all top k expressions matching a query.

PR #25, opened 2025-09-25
RFC: Blind zippers

imlvts:
As previously discussed, this moves towards Zipper API to allow "blind" zippers, where the user can track the path.
Changes:
`ZipperPath`: New trait to return path as a single continuous slice.
- `fn path` moved out of `ZipperMoving` to `ZipperPath`
- `fn move_to_path` moved out of `ZipperMoving` to `ZipperPath`. it really requires having an existing path.

`ZipperMoving`:
- `descend_indexed_byte`, `descend_indexed_branch`, `descend_first_byte`, `to_next_sibling_byte`, `to_prev_sibling_byte` now return `Option<u8>` indicating which byte we just descended to.
- `descend_until` now accepts a buffer `Option<&mut Vec<u8>>` to push the newly descended path. Could have been `Option<&mut impl Write>`, but `&mut Vec` is just simpler.
- `ascend` now returns the number of remaining steps as a `Result<(), usize>`. Maybe not the prettiest API, but forces the user to handle ascending above root. Maybe it's a good idea to separate it into a different function, and keep the original `ascend` the same.
- `ascend_until` and `ascend_until_branch` now returns `Option<usize>`, the number of bytes ascended.
- changing `to_next_step` to allow the caller to track path might require more thought, so it is left with the default implementation.


luketpeterson:
Closing in favor of https://github.com/Adam-Vandervorst/PathMap/pull/59

PR #26, opened 2025-09-25
added PolyZipper to PathMap

imlvts:
This introduces `PolyZipper`, which is an `enum` of different zipper kinds.
Each zipper method will dispatch between different zipper implementations.
This allows MORK to make queries on heterogenous sources.


imlvts:
👍

PR #33, opened 2026-04-01
zipper merge algebra (join/meet/subtract)

marcin-rzeznicki:
This PR contains zipper merge algebra (join/meet/subtract)

Implements stackless zipper-based join, meet, and difference. Adds `ZipperAlgebraExt` for ergonomic
method calls.


Adam-Vandervorst:
Except for it only being implemented on ReadZipperUntracked, LGTM. I believe the next steps are also pretty clear for lifting this out of experimental: specializing graft children, making versions available for stronger (with nodeid) and weaker (without graft) zippers.

luketpeterson:
I don't have any serious issues with this.  Sounds like it's better to land it now and deal with specific nits as we move forward.

PR #35, opened 2026-04-21
Add monomorphized N-way zipper merge with frontier-based dispatch and specialization

marcin-rzeznicki:
This PR introduces a generic N-way merge algorithm, extending the existing 2-way and 3-way implementations to arbitrary arity (well, `<= 64`) while preserving performance characteristics via specialization and careful traversal design.

#### N-way zipper merge (`zipper_merge_n_mono`)
- Fully monomorphized over `const N` and the zipper type (`SomeMutRefZ`)
- Uses a bitmask (`active`) to track participating zippers
- Stackless traversal using zipper movements and explicit depth tracking

#### Frontier-based traversal
- At each step:
- - Compute minimal byte across active zippers
- - Build a frontier (subset sharing that byte)
- - Dispatch based on frontier size:
- - - `|frontier| = 1` → graft (no descent)
- - - `1 < |frontier| < N` → subset recursion / specialization
- - - `|frontier| = N` → fast-path full descent (no recursion)
#### Specialized fast paths
- Dispatch to optimized implementations for:
- - 2-way → `zipper_merge`
- - 3-way → `zipper_merge3`
- - 4-way → `zipper_merge4`
- Larger frontiers recurse into the same function with a reduced mask

### Introduce macro-based “variadic” frontends and tuple-driven helpers for composing multiple zippers in a single call.

- New macros:
- - `zipper_meet_n!`
- - `zipper_join_n!`
- - `zipper_subtract_n!`
- New helpers:
- - `meet_n`, `join_n`, `subtract_n`

##### Syntax

Tuple-based argument passing `(&mut r1, &mut r2, ..., &rN).meet_n(&mut out)` with macro-generated impls up to 64 (current limit: 32)

or

`zipper_join_n!(r1, r2, r3, ... => w);` instead

### DIFF
--- a/src/experimental/zipper_algebra.rs
+++ b/src/experimental/zipper_algebra.rs
@@ -956,6 +956,9 @@ fn zipper_merge_n_mono<P, V, Out, A, const N: usize>(
     Out: ZipperWriting<V, A>,
 {
     debug_assert!(N > 0 && N <= 64);
+    // LLVM needs to prove: `0 ≤ i < N` But i comes from: `i = bits.trailing_zeros() as usize;`` So the
+    // compiler must connect: “this bitmask only contains bits < N”
+    assert!(active >> N == 0);
 
     #[inline]
     fn active_bits<const N: usize>(active: u64) -> impl Iterator<Item = usize> {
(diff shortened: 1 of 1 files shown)


src/experimental/zipper_algebra.rs
```diff
@@ -543,7 +607,7 @@ where
                         cmp_swap(&mut b, &mut c);
                         if let Some(next) = b {
                             P::on_single(lhs, L as u64, ByteMask::from_range(min..next), out);
-                            lhs_idx = lhs_mask.index_of(next)
+                            l = (lhs_mask & ByteMask::from_range(next..)).next_bit(0);
```
marcin-rzeznicki, line 610: Perhaps move to `ByteMaskIter` or `ByteMask`

luketpeterson (approved):
This is good.

Marcin will move the ByteMaskIter extra functionality into the utils after this call.

PR #36, opened 2026-04-29
zipper: short-circuit on shared subtries

marcin-rzeznicki:
Enhance merge logic by detecting structural sharing:
- on function entry
- immediately after descending on a shared edge

If all active zippers share the same node, delegate to `MergePolicy::on_id`
and skip the entire subtree.

### DIFF
--- a/src/experimental/zipper_algebra.rs
+++ b/src/experimental/zipper_algebra.rs
@@ -55,13 +55,13 @@ pub use zipper_algebra_poly::ZipperMergeF;
 /// - [`zipper_meet`]
 /// - [`zipper_subtract`]
 pub trait ZipperAlgebraExt<V: Clone + Send + Sync, A: Allocator = GlobalAlloc>:
-    ZipperInfallibleSubtries<V, A> + ZipperMoving + Sized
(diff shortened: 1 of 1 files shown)


Adam-Vandervorst:
Since this code is contingent on a criterion that does occur in PathMaps loaded from a set (without merkleize), I'd like to see some tests that specifically exercise this codepath.

marcin-rzeznicki:
> Since this code is contingent on a criterion that does occur in PathMaps loaded from a set (without merkleize), I'd like to see some tests that specifically exercise this codepath.

Fixed in 9ff3982 - I think the test gives us enough coverage

Adam-Vandervorst:
Yes, at least this is some sniff test, and we can ramp up the testing with the property-based testing when we move this out of experimental.

Adam-Vandervorst:
I'm good with merging it now.

marcin-rzeznicki:
Thanks! @luketpeterson do you have any remarks?

luketpeterson:
Marcin explained the thinking and it's probably the best way.  Merging.

PR #37, opened 2026-05-08
zipper_algebra: support more zippers

marcin-rzeznicki:
A small PR that just extends `ZipperAlgebraExt` ( `z.join()` syntax ) and `SomeMutRefZ` ( `zipper_join_n!(...)` and `(z1, z2,...).join` ) support to more zippers. Concretely, `ZipperAlgebraExt` is 'implemented' for `ReadZipperOwned` and `PrefixZipper`. `SomeMutRefZ` now covers `PrefixZipper` for `ReadZipperUntracked` and `ReadZipperTracked`.

Note: There are some limitations: `SomeMutRefZ` cannot support `ReadZipperOwned` because it forces `V: 'static` which we don't want; it also cannot support arbitrary `PrefixZipper` because it can't be expressed without GADTs (or, at least I don't know how)

PR #41, opened 2026-06-03
zipper_algebra: support more zippers

marcin-rzeznicki:
Copy of #37


marcin-rzeznicki:
Merging into @luketpeterson branch, as discussed on a call

PR #42, opened 2026-06-12
zipper_algebra: Add XOR

marcin-rzeznicki:
This change introduces an n-way XOR operation implemented on top of the
existing ordered zipper traversal framework.

The implementation treats tries as sparse path→value mappings, with missing
paths interpreted as the lattice bottom element. XOR is therefore computed
pointwise on values rather than on path presence:

```
result(p) = xor(v₁(p), ..., vₙ(p))
```

A deliberate design choice is that traversal follows join semantics: a child
edge is explored whenever it is present in at least one input. Although XOR is
often described in terms of parity, cancelling values at a path does not imply
that all descendants cancel as well. Stopping traversal based solely on path
parity would therefore miss surviving values in deeper subtries.

As a result, XOR shares the same traversal skeleton as join and differs only in
the algebra used to combine coincident values.

### DIFF
--- a/src/experimental/zipper_algebra.rs
+++ b/src/experimental/zipper_algebra.rs
@@ -1690,9 +1690,358 @@ where
     }
 }
 
+/// A conjunction clause in a Disjunctive Normal Form (DNF) expression.
+///
+/// A clause is represented as a bitmask over a fixed universe of `N` input
+/// zippers. Bit `i` is set iff zipper `i` participates in the conjunction.
+///
+/// For example, for `N = 4`:
+///
+/// ```text
+/// {0,2}  => 0b0101
+/// {1,3}  => 0b1010
+/// ```
+///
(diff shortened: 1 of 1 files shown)


Adam-Vandervorst:
Let's go from `xor` to `sym_diff`. And I guess `xor_n` becomes `odd_occ`?

marcin-rzeznicki:
> ` [42e0dea](/Adam-Vandervorst/PathMap/pull/42/commits/42e0dea29e12994d8e4b7fdcd73f24cea3817063)`

Fixed in [42e0dea](https://github.com/Adam-Vandervorst/PathMap/pull/42/commits/42e0dea29e12994d8e4b7fdcd73f24cea3817063). I opted for `sym_diff` to be used consistently, rather than `sym_diff` for some args and `odd_occ` for the others.

PR #45, opened 2026-06-23
Hardening: dense meet operand fix, a stacked-borrows UB fix, and two guards

MesTTo:
## Summary

Five small, independent correctness commits, each proven against a stock
`master` reproduction before the fix:

- **Fix composite meet/join operand selection in DenseByteNode** — the
  differential added in this commit fails 2/4 on stock `master`
  (`combine_algebraic_results` was copying `self` unconditionally in a case
  where the counter side held the correct operand); with the fix it's 4/4.
- **Fix stacked-borrows UB in `zipper_merge_n_mono` via safe
  `get_disjoint_mut`** — miri (stacked borrows) aborts on stock `master` at
  the raw-pointer loop in `with_k`; replacing it with the stable (≥1.86)
  `get_disjoint_mut` is miri-clean.
- **Reject malformed serialized offsets** — stock `master` panics on
  deliberately corrupted offset bytes during deserialization; this converts
  the failure into a proper `Err`.
- **Guard line-list node layout** — a compile-time size assert pins
  `LineListNode<[u8;1024]>` at 64 bytes under `slim_ptrs` on x86_64. Only that
  configuration is asserted: building `not(slim_ptrs)` currently fails (26
  errors, one root cause — `TrieNodeODRc`'s Arc-backed variant has no
  empty-sentinel representation, so `new_empty`/`is_empty`/`make_unique`/`==`
  don't exist there). That's the open question the `TaggedNodeRefMut::EmptyNode`
  GOAT note already flags; happy to take a pass at it once there's a
  preferred sentinel design.
- **Re-enable the `tree_serde_2` test** — verified passing on this tree; the
  `#[ignore]` was stale.

## Test plan
- [x] `cargo test --release` (663/0/0, up from the 660/0/1 baseline)
- [x] `cargo test --release --test pathmap_algebra_differential` (4/4)
- [x] `cargo miri test --release` over the touched zipper_algebra tests (120/120)
- [x] `cargo test --release --doc` (7/0)


adamv-symbolica:
The changes to LineListNode seem to hint at a bug-fix, can you give an explanation of the bug? The test for it is quite impenetrable to me.

MesTTo:
The LineListNode change is not fixing a newly observed runtime failure by itself... It is guarding an existing layout invariant... LineListNode is written around a fixed in-memory layout. The key byte capacity is chosen so the node stays at 64 bytes with `slim_ptrs` on x86_64, and 48 bytes without `slim_ptrs` because of the DynBox header. If that size changes, the node can stop matching the assumptions behind `KEY_BYTES_CNT` and the cache-line/list-node packing. The old test only checked this at runtime when that test was selected; the new const assertion makes layout drift a compile-time failure. the test is not meant to prove a semantic transformation; it is just the serialised form of the malformed-offset case

src/dense_byte_node.rs
```diff
@@ -1864,7 +1873,12 @@ impl<V: Clone + Send + Sync, A: Allocator, Cf: CoFree<V=V, A=A>, OtherCf: CoFree
                 if new_mask > 0 {
                     AlgebraicResult::Identity(new_mask)
                 } else {
-                    AlgebraicResult::Element(Self::new(self.rec().cloned(), None))
+                    let rec = if rec_mask & SELF_IDENT > 0 {
+                        self.rec().cloned()
+                    } else {
+                        other.rec().cloned()
+                    };
+                    AlgebraicResult::Element(Self::new(rec, None))
```
adamv-symbolica, line 1986: Sorry I meant DenseNode -- i.e. the motivation behind these.

MesTTo:
ah right, denseNode not linelist... that's the actual fix, sorry.

it's in `combine_algebraic_results`. when you combine two cofrees and one half (rec or val) comes back as `Identity(mask)`, the mask says which operand the result equals: `SELF_IDENT` for self, `COUNTER_IDENT` for other. once the combined cofree isn't identical to either operand (`new_mask == 0`) it builds a fresh `Element`, and for the surviving half the old code just cloned `self.val()` / `self.rec()`.

that's only right when the surviving half is identity-to-self. in a commutative op the half can be identity-to-other instead (SELF_IDENT cleared, COUNTER_IDENT set), and then cloning self's half drops the wrong operand's value/child into the node. so the fix reads the bit: `mask & SELF_IDENT > 0` takes self, else takes other. same for the rec arm.

the giveaway is the `(rec_el, val_el)` arm right below... it already picks the operand by index (`0 => self`, `1 => other`), the two identity-collapse arms just weren't doing the same. this makes them consistent.

the line 698 hunk in `node_get_payloads` is the same accounting on the read side: a 1-byte value request landing on a cofree that still has a rec link wasn't flagging that the rec half is there.

no change for the self-identity case, it only diverges when the surviving half matched other.

luketpeterson:
It looks like the ByteNode issues were serious, so good work fixing those.

I tweaked the tests to work (and finish in time) under miri.  But that wasn't structural.

I want @marcin-rzeznicki to weigh in on the changes to zipper_algebra.rs as they have a potential to affect performance.

MesTTo:
the only hunk in `zipper_algebra.rs` is `with_k`, and it is on the merge inner loop, so the concern is
fair. i measured it rather than argue it. all of this is against current master (0fff681), not the
PR's recorded base, since master moved 22 commits under this branch. `with_k` itself is unchanged
between the two.

first the bug, which is real, and i did not have to write a test to catch it because yours already
does. i ran miri under its default stacked borrows over
`experimental::zipper_algebra::tests::join::*`:

```
with_k as on master   test_almost_identical_paths_n ... error: Undefined Behavior:
                      trying to retag from <...> for Unique permission, but that tag
                      does not exist in the borrow stack for this location
with_k in this PR     26 passed; 0 failed
```

i traced it to the `xs.as_mut_ptr()` inside the loop: it is a `&mut self` call, so each iteration
retags the whole slice as Unique and invalidates the pointer the previous iteration took. one caveat
i should state up front, since it cuts against me: i also ran it under tree borrows and all three
versions pass there, so this is a stacked-borrows violation and not a tree-borrows one.

then the cost. i ran callgrind for instructions retired over a fixed workload driving `zipper_n_join`
and `zipper_n_subtract` at N=6 and N=4, with all builds producing an identical checksum:

```
with_k as on master        815,709,825
with_k in this PR          810,993,408   (-0.578%)
```

the whole program retires fewer instructions with the PR applied, but i do not want to sell that as a
win: the difference is inlining churn (a `drop_glue` clone moves between symbols), not the change
paying for itself. the honest read is no measurable regression.

`with_k` itself does get more expensive, exactly as you would expect. when i sum the delta over the
`zipper_merge_n_mono` monomorphizations i get +1,399,257 instructions over 88,155 `with_k` calls,
about 16 instructions per call. i attribute that to `get_disjoint_check_valid`, which std documents
as `N*(N+1)/2` comparisons, so 3, 6 and 10 for K=2, 3, 4. the indices come out of
`bits.trailing_zeros()`, so llvm cannot prove distinctness and cannot fold the check away.

if you think 16 instructions per call is 16 too many, i have a version with no check at all that
still fixes the UB. hoisting the one `as_mut_ptr()` above the loop removes the re-borrow, which was
the entire problem:

```rust
let base = xs.as_mut_ptr();
let mut ptrs: [*mut T; K] = [std::ptr::null_mut(); K];
let mut i = 0;
while i < K {
    let idx = bits.trailing_zeros() as usize;
    bits &= bits - 1;
    ptrs[i] = unsafe { base.add(idx) };
    i += 1;
}
let refs = unsafe { ptrs.map(|p| &mut *p) };
```

i measured that one at 815,710,722 instructions against master's 815,709,825, a difference of 897 in
815 million, and it passes the same 26 miri tests. it keeps `unsafe`, which is why i did not propose
it first, but it is strictly zero-cost and the safety argument becomes one sentence instead of three.

my own read is that the PR should go in as written: 16 instructions on a call that is already doing a
trie descent is not a cost worth carrying `unsafe` for, and the measurement says the suite does not
notice. i am offering the hoist because it is cheap to swap in, not because i think the current
version needs it. if @marcin-rzeznicki sees a workload where the check does show up, that is the
version to take.

one thing that made this harder to answer than it should have been: i could not find anything in
`benches/` that reaches this code. `zipper_n_*` and `zipper_merge_dnf` are behind `zipper_alg` and
have no benchmark, so a regression here would not show up in CI. i wrote one to get the numbers
above and can push it with this PR if you want it. it drives N=6 and N=4 join/subtract over
prefix-heavy tries and hits `with_k` 88,155 times per run at K=2/3/4. while writing it i noticed
`meet` never reaches `with_k` at all, since `Meet::descend_on_some_equal` is false and a partial
frontier annihilates instead of descending, so meet is unaffected by any of this.

i also re-ran the suite: 120/120 `zipper_algebra` tests pass on all three versions. and thanks for
the miri and environment fixes on the serialization and differential tests, those were mine to get
right.

src/experimental/zipper_algebra.rs
```diff
@@ -1423,23 +1423,23 @@ fn with_k<const K: usize, T, R>(
...
-
-    // SAFETY:
-    // - indices are distinct (bitmask)
-    // - derived from same slice
-
-    // should be zero-cost after inlining
-    let refs = unsafe { ptrs.map(|p| &mut *p) };
+    let refs = xs
+        .get_disjoint_mut(indices)
+        .expect("active bitmask indices are distinct and in bounds");
```
Adam-Vandervorst: We can't introduce runtime checks for compile time properties.

luketpeterson: get_disjoint_unchecked_mut has no bounds or overlap-checking cost, so using that should make the runtime cost equivalent, and will keep miri happy.

luketpeterson: Changed.

MesTTo:
@Adam-Vandervorst on "we can't introduce runtime checks for compile time properties": fair, and
@luketpeterson's `get_disjoint_unchecked_mut` is the right resolution. i measured it rather than
assume, same harness and workload as my numbers above, callgrind instructions on a fixed workload
driving `zipper_n_join` and `zipper_n_subtract` at N=6 and N=4, all four builds producing an
identical checksum:

```
with_k as on master                 815,711,000
get_disjoint_mut (what i pushed)    810,993,698
hoisted base pointer               815,710,008
get_disjoint_unchecked_mut (now)   809,885,377
```

so both of luke's claims hold. the check was costing 1,108,321 instructions, which over 88,155
`with_k` calls is about 12.6 each, and that matches std documenting `get_disjoint_check_valid` as
`N*(N+1)/2` comparisons, 3/6/10 at K=2/3/4. and it stays miri-clean: 26/26 of
`experimental::zipper_algebra::tests::join::*` pass under stacked borrows, and under tree borrows.
the unchecked version is the cheapest of the four, so this is strictly better than what i had.

i also checked the obligation it now carries rather than taking it on faith. distinctness comes from
the bitmask, and in-bounds holds because `Clause::from_mask` asserts `mask >> N == 0` and
`zipper_merge_n_mono` asserts `N > 0 && N <= 64`, so the indices cannot leave the array. it is sound.

one thing i would still add: the `// SAFETY:` comment went with the rewrite, and `with_k`'s own
`debug_assert!` only covers the popcount, not the range. so nothing local says why the unchecked call
is allowed, and the two facts that make it allowed are in `Clause` and in the caller. a restored
SAFETY line plus a range `debug_assert!` costs nothing in release and means the next person changing
the frontier construction gets told. happy to push that if you want it.

on the rest of the PR: the ByteNode/DenseNode fix and the layout guard are unchanged from the
explanation above, and the 120/120 `zipper_algebra` tests pass on every variant i tried.

Adam-Vandervorst (approved):
LGTM.

luketpeterson:
A SAFETY comment wouldn't hurt.  Rust style says every unsafe block should
have them.  It's a rule we don't consistently follow but I'm in favor.

Just make sure the comment is short and straight to the point.

Also, you probably know this already, but the unchecked variants still
perform the checks in debug builds.

PR #46, opened 2026-06-23
Add runtime counter snapshots

MesTTo:
Adds the counter API only.

The branch introduces runtime counter snapshot, reset, read, and record helpers behind the existing counters feature. It does not wire counters into unrelated call sites.

Validation:

- `cargo check --features counters`
- `cargo test --features counters counters::tests`
- `jscpd` on `src/counters.rs`

### DIFF
--- a/src/counters.rs
+++ b/src/counters.rs
@@ -1,6 +1,432 @@
 use crate::PathMap;
-use crate::zipper::{*, zipper_priv::ZipperPriv};
+use crate::zipper::*;
 use crate::trie_node::{TaggedNodeRef, TrieNode};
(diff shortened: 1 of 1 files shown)


Adam-Vandervorst:
I don't see the value of this in its current form -- is this still WIP? Where would it be used?

MesTTo:
yeah you're right... i checked and nothing in pathmap actually touches these. the record sites and the readers both have zero callers, so on its own it's just dead plumbing. no point landing it like this. closing it and i'll bring the counters back together with the code that reads them. thanks for flagging.

PR #47, opened 2026-06-23
Skip dense meet joined-mask scans

MesTTo:
## Summary

Implements the TODO in the dense meet: instead of walking both dense nodes'
full masks and re-testing membership per byte, walk the intersection mask
directly, so bytes absent from either side are never visited. No public API
change; `is_identity` handling is equivalent (the intersection walk observes
the same element/identity outcomes the full walk did).

## Result

Measured on this branch vs `master` (interleaved, two reps each side, best
median): `superdense_meet` 8000 26.1→25.0us, 16000 42.5→39.8us, 32000
100→93.9us (a consistent 4-6% where the masks are dense); `sparse_meet` and
`binary_meet` at parity to -2% (their masks are thin, so there's little scan
to skip). Composes cleanly with `pr/pathmap-hardening`'s DenseByteNode
operand-selection fix (both touch this file; they auto-merge, and the
algebra differential passes with both applied).

## Test plan
- [x] `cargo test --release --lib` (660/0, matches the `master` baseline)
- [x] `cargo test --release --test pathmap_algebra_differential` (4/4)
- [x] `cargo bench --bench superdense_keys / sparse_keys / binary_keys -- meet`

### DIFF
--- a/benches/superdense_keys.rs
+++ b/benches/superdense_keys.rs
@@ -160,6 +160,26 @@ fn superdense_meet(bencher: Bencher, n: u64) {
     });
 }
 
+/// This is the success case for dense meet's intersection-mask walk: both
+/// nodes have many children, but only a small fraction of those children are
+/// present in both nodes.
+#[divan::bench(args = [1000, 2000, 4000, 8000, 16000, 32000])]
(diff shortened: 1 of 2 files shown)


luketpeterson:
This seems to be a good change.  I added an extra benchmark to illustrate the gain.

PR #48, opened 2026-06-23
Add zipper-type property tests: same_trie, seeded ring algebra, long-path serialization

MesTTo:
## Summary

Three additive test/API commits, no internals changed:

- **`PathMap::same_trie`** — an O(1) copy-on-write root-identity staleness
  check (compares the root node pointer, no traversal).
- **Seeded ring algebra property tests** — implements the fuzz test the
  trailing TODO in `ring.rs` described: seeded pseudorandom HashSet and
  nested-HashMap lattices checked against plain set oracles across
  `pjoin`/`pmeet`/`psubtract`/`join_into`, asserting both results and the
  identity/element status masks. Also covers the empty-subset-collapse case
  and the HashMap `psubtract` case the inline notes flagged as untested. Only
  the now-resolved TODO notes are removed; the still-open ones (`Vec`
  `SetLattice`, `LatticeCounter`/`LatticeBitfield` traits) are left in place.
- **Long-path serialization coverage** — answers the note at the top of
  `paths_serialization.rs`: round-trips deterministic paths of 7 / 2049 /
  4097 / 24000 bytes, so the input spans the internal 4096-byte compression
  chunk, checked in both directions.

## Test plan
- [x] `cargo test --release --lib` (664/0, up from the 660/0/1 baseline)
- [x] `cargo test --release --lib same_trie` / `seeded` / `paths_serialization` individually


src/ring.rs
```diff
@@ -1332,14 +1623,17 @@ mod tests {
         inner_map_1.insert("1", ());
         a.0.insert("A", inner_map_1.clone());
         b.0.insert("B", inner_map_1);
-        // b.0.insert("C", HashMap::new()); TODO: We might want to test collapse of empty items using the is_bottom() method
```
Adam-Vandervorst, line 1335: This comment is deleted but is_bottom check is not introduced?

src/trie_map.rs
```diff
@@ -123,6 +123,25 @@ impl<V: Clone + Send + Sync + Unpin> PathMap<V, GlobalAlloc> {
...
 impl<V: Clone + Send + Sync + Unpin, A: Allocator> PathMap<V, A> {
+    /// Whether `self` and `other` share the same root node by pointer (O(1)).
+    /// PathMap is copy-on-write, so a map and a clone of it stay `same_trie`
+    /// until one is mutated: any insert or removal replaces the root node with a
+    /// fresh allocation. A `true` result therefore proves the trie's node
+    /// structure has not changed relative to `other`. Because `other` (typically
+    /// a cheap COW clone kept as a snapshot) holds the old root alive, its
+    /// address cannot be reused while the snapshot exists, so this is free of the
+    /// ABA pointer-reuse hazard. Root VALUES (the empty-key entry) are not
+    /// compared; callers that key a cache on subtree contents do not touch it.
```
Adam-Vandervorst: This seems like a verbose and at the same time unspecific explanation.

src/trie_map.rs
```diff
@@ -123,6 +123,25 @@ impl<V: Clone + Send + Sync + Unpin> PathMap<V, GlobalAlloc> {
...
+    #[inline]
+    pub fn same_trie(&self, other: &Self) -> bool {
```
Adam-Vandervorst: @luketpeterson good public interface addition? To me, "same trie" could also be logical, so I'd change this to "ptr eq".

Adam-Vandervorst (changes requested):

src/trie_map.rs
Adam-Vandervorst: Should also use node_id?

MesTTo:
pushed a fix. #4 was the important one, it's a real bug.

renamed `same_trie` to `root_ptr_eq`. your read on the name was right, and it's wrong the other way too, it's false for separately-built identical tries. rewrote the doc against `shared_node_id`'s shape.

the bug #4 points at: it compared only the root node, but root values live in `root_val` outside the node, and `set_val_at` on an empty key goes through `set_root_val` without touching `root`. so two maps differing only by a root value compared equal, i.e. true for different contents. added `root_ptr_eq_is_false_when_a_root_value_differs`, which fails on the old code. fixed by returning false when either map has a root value, so a true result means same contents. same reason `shared_node_id` returns None at a node with a value.

on the ring.rs one: the `is_bottom` that TODO wanted was a `Lattice` method against `bottom()`, and `bottom()` was deleted from the trait (04732ef), so it never landed. the empty-collapse it wanted tested is real though, it's `result_set.len() == 0 -> AlgebraicResult::None` at ring.rs:1070, which is what the test exercises.

src/trie_map.rs
luketpeterson: I ended up implementing ZipperConcrete instead.  It provides more useful API capability and I think it's more consistent.

@MesTTo WDYT?

src/ring.rs
luketpeterson: The is_bottom method is also deleted.

luketpeterson (approved):

MesTTo:
**on `ZipperConcrete`:** agreed, it should replace `root_ptr_eq`. it gives callers something
`root_ptr_eq` could not, and the thing i actually cared about survived the swap:
`self.root_val().is_some()` in `PathMap::shared_node_id` is the bug my test was pinning. root values
live outside the root node, so a node-level identity check cannot see them, and without that guard
two maps differing only by a root value report identical.

i went looking for whether i could have got there through a zipper instead, and you could not have:
on master the zipper route returns `None` at every root, because `ReadZipperCore::is_shared`
hardcodes `false` there (`// root`), `TrieRefBorrowed`/`TrieRefOwned::is_shared` return `false`
unconditionally, and `read_zipper_shared_node_id` short-circuits on `!is_shared()`. so putting it on
`PathMap` was the only place it could go, and before your commit i do not think there was any call
sequence that answered this question.

two small asks, neither blocking:

1. the new impl has no doc comment, and the contract is not obvious. `Some(x) == Some(y)` means same
   root node *and* no root value on either side, therefore same contents. `None` means unknown, not
   different. that asymmetry is the whole reason the `root_val` guard is there, and the trait docs
   say the opposite thing about `is_shared` ("must never rely on the return value for correctness"),
   which the new impl does rely on. worth writing down for whoever touches it next.
2. the new test asserts `with_root_value.shared_node_id() == None` on one map. the regression it
   replaces asserted the predicate on a *pair* differing only by a root value. one line restores that
   shape: `assert_ne!(m.shared_node_id(), with_root_value.shared_node_id())` and the reverse.

one observation rather than a request, which i noticed while checking the above: the two impls now
disagree about the same location. for a map whose root is shared with a clone, `PathMap::is_shared()`
is `true` and gives `Some(id)`, while `read_zipper().is_shared()` is `false` and gives `None`. i
think yours is the useful answer and the zipper's `false // root` is the artifact, so i read it as a
question about whether the zipper root case should be fixed to match, not as anything this PR needs
to resolve.

i am also giving up some precision by taking your version, all of it in the safe direction. i now get
`None` where `root_ptr_eq` returned `true`, for two empty maps, for an unshared map against itself,
and for two identical maps that both carry a root value. none of that touches the cow-staleness case
i wrote it for, since a snapshot clone always drives the refcount to 2, so i am fine losing it. the
one new hazard i would watch is that `shared_node_id` hands back a storable `u64`, so caching it
across a drop can collide with a later allocation, where `root_ptr_eq` took two live references and
could not. the trait doc already warns about it.

**on `is_bottom`:** right, both are gone. i grepped and the only occurrence left in the tree is the
TODO comment itself. `bottom()` went in 04732ef ("Cleaning up a bunch of dead code"), and it was
already commented out before that, together with the note proposing `is_bottom`. so the check that
TODO wanted was never implemented and the trait method it would have compared against is gone too.

the behaviour it wanted tested is real though, and i traced it to
`set_lattice_integrate_into_result`: `result_set.len() == 0` returns `AlgebraicResult::None`, and the
caller drops it so the key never enters the outer set. the assertion five lines below the TODO,
`assert!(joined.get(&"C").is_none())`, was vacuous on master, since neither operand had `"C"`. this
PR inserts `"C"` as an empty sub-map into both, so the inner `pjoin` of two empty maps actually runs
and only the collapse keeps it out.

i checked that rather than assuming it. i patched `if result_len == 0` to `if false` and re-ran:
`set_lattice_join_test2` still passes on master, and fails on this branch with `left: 3, right: 2`.
so the test does exercise the path, and it is the insert that makes it do so.

i will also fix the `unused import: AlgebraicStatus` in ring.rs while i am in here.

luketpeterson:
> on master the zipper route returns None at every root, because ReadZipperCore::is_shared hardcodes false there

The reason it was hardcoded to false was a leftover from a limitation that no longer exists.  Pushed the fix for that.  But leaving the ZipperConcrete impl on PathMap because it saves the overhead of creating a zipper.

luketpeterson:
@MesTTo Do you plan to push more code changes?  Your last comment seemed to say you were planning further changes.

MesTTo:
pushed two commits, and that is all i had planned, so this is ready from my side.

- dropped the unused `AlgebraicStatus` import in the ring tests. it was the only warning in a
  `--tests` build.
- turned the root-value case into a pairwise assertion.

on the second one: your `assert_eq!(with_root_value.shared_node_id(), None)` already catches a
missing guard, so this is not filling a hole. what it adds is the shape the regression actually had.
the old bug was not "a map with a root value answers wrongly", it was "these two maps report the
same identity while holding different contents", and only a pair can say that. i checked it fires on
its own rather than assuming: with the `root_val` guard removed and your isolated assertion
neutralised so it cannot mask mine, the new one fails with

```
assertion `left != right` failed
  left: Some(101365311012208)
 right: Some(101365311012208)
```

which is the old behaviour exactly.

666 lib tests pass and the `--tests` build is warning-free.

thanks for fixing the zipper root case rather than leaving it. and agreed on keeping both impls,
skipping the zipper construction is a better reason than the one i had for putting it on `PathMap`.

luketpeterson:
Ok.  Merging.

src/trie_map.rs
MesTTo: answered this in the thread on the PR itself, but leaving it here so this one is not left hanging:
agreed, `ZipperConcrete` is the better call and i would not keep `root_ptr_eq` alongside it.

the reason i am comfortable is that the guard i actually cared about survived the swap:
`self.root_val().is_some()` in `PathMap::shared_node_id`. root values live outside the root node and
`set_root_val` does not touch `root`, so without it two maps differing only by a root value report
the same identity. i pushed the pairwise assertion for that case and checked it fires on its own with
the guard removed.

the precision i give up going from `root_ptr_eq` to `Some(x) == Some(y)` is `None` where the old one
said `true`, for two empty maps, an unshared map against itself, and two identical maps that both
carry a root value. none of that touches the cow-staleness case it was written for, so it is a fine
trade.

PR #50, opened 2026-07-04
Add a zipper_head_owned benchmark

MesTTo:
## Summary

Adds a divan benchmark comparing an owned `ZipperHead` read-zipper creation
against the existing borrowed path, answering the TODO left in
`src/zipper_head.rs` about benchmarking that difference.

## Result

Owned creation runs at roughly parity with the borrowed path (~150ns vs
~140ns per creation in local runs) — the GOAT blocks describing the open
design questions are left untouched; this only adds the measurement they
were missing.

## Test plan
- [x] `cargo test --release` (660/0, matches origin/master's baseline)
- [x] `cargo bench --bench zipper_head_owned`

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -128,6 +128,10 @@ harness = false
 name = "graft_masked_branches"
 harness = false
 
+[[bench]]
+name = "zipper_head_owned"
+harness = false
+
 [workspace]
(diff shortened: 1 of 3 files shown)


Adam-Vandervorst:
Ty!

PR #51, opened 2026-07-04
Add copy-on-write write-path counters

MesTTo:
## Summary

A follow-up to #46, which you correctly flagged as dead plumbing (the
record sites and readers had zero callers). This time the writer and reader
are both wired to real code:

- **Writer**: two lines in `TrieNodeODRc::make_unique` — the one choke point
  every structural write in the library goes through — record whether the
  call found the node already unique or had to pay the copy-on-write clone.
  Compiled out entirely without the `counters` feature; the default build is
  byte-for-byte unchanged.
- **Reader**: `cow_counters()` / `reset_cow_counters()` return a
  `CowCounters { make_unique_calls, cow_clones }` snapshot, exercised by an
  integration test in its own process: writing an unshared trie records zero
  clones, writing while another handle aliases it records at least one clone
  (and never more clones than calls), the aliased handle is unaffected, and
  writes after the alias is dropped stop cloning.

Also includes a standalone fix: the `counters` feature didn't compile on
`master` (an import of a `zipper_priv::ZipperPriv` that no longer exists —
`get_focus` moved to the public `ZipperInfallibleSubtries` — plus a doc
example referencing an undefined `map` and leaving a generic parameter
uninferred). Fixed as its own commit before adding the new counters.

## Test plan
- [x] `cargo test --release --features counters` (660/0 lib, 9/0 doc)
- [x] `cargo test --release --features counters --test cow_counters` (1/1)
- [x] `cargo test --release` with the feature off (660/0, unchanged)


Adam-Vandervorst:
What's the conclusion of this work? What is the projected future use?

MesTTo:
conclusion: CoW is attributable at exactly one place, `make_unique`, which every structural write goes through. the test pins the contract, writing an unshared trie never clones (`cow_clones` 0), writing an aliased one clones at least one node and never more than the `make_unique` calls, the alias stays untouched, and dropping it stops the cloning.

projected use: it's the instrument for the write-zipper work where the goal is to cut cloning, so an optimization can assert "this movement/read didn't clone" on the counter instead of reading a timer. unlike #46, `make_unique` records and the test reads, so it's not dead plumbing.

the feature-build fix stands alone too, `--features counters` didn't compile on master (stale `zipper_priv::ZipperPriv` import).

luketpeterson:
Can you say more about the idea / use case behind this feature?

Is this for performance optimization for people working on pathmap itself?  Or is it for use by downstream code?

luketpeterson:
@MesTTo See question above.

MesTTo:
the first one. i wrote it for people working on pathmap itself, which lines up with what the feature
already says it is for in Cargo.toml ("Mainly useful when tuning performance of the PathMap
internals").

concretely, what i was after: copy-on-write is attributable at exactly one place, `make_unique`,
which every structural write goes through. counting there lets me turn "did this change avoid
cloning" into an assertion instead of a wall-clock comparison. that is what i want it for in the
write-zipper work, where my goal is to cut cloning: i can assert `cow_clones == 0` for a movement or
read that should not clone, and that holds on any machine and under any load. a timer cannot tell me
that.

so i wrote the test to pin the contract rather than the implementation: writing an unshared trie
never clones, writing an aliased one clones at least one node and never more than the number of
`make_unique` calls, the alias stays untouched, and dropping the alias stops the cloning.

downstream use is possible but is not why i wrote it, and i would not design the API around it. the
counter is a global under a non-default feature, so i think of it as an instrument for a build you
are already profiling, not something to condition behaviour on.

separately, the second commit stands on its own regardless of what you decide about the counters.
`--features counters` does not compile on master today. i just re-checked on 0fff681:

```
error[E0432]: unresolved import `crate::zipper::zipper_priv::ZipperPriv`
error: could not compile `pathmap` (lib) due to 1 previous error
```

i am happy to split that into its own PR if you would rather take the fix and think about the
counters separately. one last thing, since i think it was the fair objection to #46: there
`make_unique` recorded and nothing read it, whereas here the test reads it, so i am not adding dead
plumbing again.

luketpeterson:
I see that can be useful.  I'll merge it.  Thanks!

PR #53, opened 2026-07-07
Add ArenaCompactTree::merge_zipper_into_file: append-only merge into an ACT file

imlvts:
Merges any zipper (Zipper + ZipperMoving + ZipperValues) into an existing on-disk ACT by appending only the new data and rewriting the root offset; existing bytes are never modified. Backward relative offsets let appended nodes reference old subtrees directly, so untouched subtrees are shared byte-for-byte. Because siblings must be stored contiguously, a node with any changed child gets its sibling run re-emitted as shallow copies whose children still point into the old arena; line nodes are split as needed, reusing the old line data when a whole segment survives. On value conflicts the zipper's value wins, and a merge that adds nothing leaves the file untouched.

Works with PathMap read zippers and ACT zippers alike (ACT -> ACT merge), enabling incremental accumulation into one ACT file without rebuilding it.

Also fixes get_val_at reading past a branch's last sibling when asked for a byte absent from the child mask (missing test_bit check; subtract-overflow panic in debug builds).


src/arena_compact.rs
```diff
@@ -732,6 +732,9 @@ where Storage: AsRef<[u8]>
                     if path.is_empty() {
                         return node.value;
                     }
+                    if !node.bytemask.test_bit(path[0]) {
+                        return None;
+                    }
```
Adam-Vandervorst, line 781: Do you have a regression case for this?

src/arena_compact.rs
```diff
@@ -2257,6 +2693,166 @@ mod tests {
...
+        }
+        assert_eq!(merged.get_val_at([0, 0]), None);
+    }
+
+    #[test]
+    fn test_act_merge_repeated_and_act_source() {
+        // Several merge waves over an LCG key soup; the last wave merges from
+        // an ACT zipper instead of a PathMap zipper.
+        let dir = tempfile::tempdir().unwrap();
+        let file = dir.path().join("waves.act");
```
Adam-Vandervorst: waves.act?

Adam-Vandervorst (approved):

src/arena_compact.rs
imlvts: This actually fixes a bug. prolly need to add a test case to showcase the bug.

src/arena_compact.rs
imlvts: The name doesn't matter much. It's a collection of randomly generated sequences.

Adam-Vandervorst: Maybe useful in other places; can keep it in the repo?

src/arena_compact.rs
Adam-Vandervorst: Yes

src/arena_compact.rs
imlvts: (done)

src/arena_compact.rs
imlvts: (done)

PR #56, opened 2026-07-28
Making DependentProductZipperG behave as a compliant ProductZipper

luketpeterson:
Applying ProductZipper tests to DependentProductZipperG
Expanding ProductZipper tests
Fixing test failures in DependentProductZipperG

### DIFF
--- a/src/dependent_zipper.rs
+++ b/src/dependent_zipper.rs
@@ -61,19 +61,14 @@ impl<'trie, PrimaryZ, SecondaryZ, V, C, F : Clone + for <'a> FnOnce(C, &'a [u8],
     /// Actual focus factor calculation.
     /// Returns a valid index into `self.factor_paths`, truncating to parents if requested.
     fn factor_idx(&self, truncate_up: bool) -> Option<usize> {
+        debug_assert_eq!(self.factor_paths.len(), self.secondary.len(),
+            "factor_paths and secondary must stay in step");
         let len = self.path().len();
         let mut factor = self.factor_paths.len().checked_sub(1)?;
         while truncate_up && self.factor_paths[factor] == len {
             factor = factor.checked_sub(1)?;
         }
         Some(factor)
-    }
-
-    /// Returns the number of factors composing the `DependentProductZipperG`
-    ///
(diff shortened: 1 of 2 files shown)


luketpeterson:
@Adam-Vandervorst There were some issues where DependentProductZipperG appeared to be broken.  But I don't understand all the situations in which you use it, so I didn't want to make changes that could potentially break behavior without giving you a chance to weigh in.

Adam-Vandervorst:
The implementation was copy-pasted from the product zipper, so probably just an out-of-sync problem with a fix that wasn't applied to both. I'll run the MORK tests, it's used quite a bit.

luketpeterson:
No.  That was my thought too, at first.  But it's actually a mismatch
between the product zipper contract and the way it was adapted into the
dependent product zipper.  I wouldn't have bothered with the pr if it was
just some fixed hitting ProductZipperG but not being applied here.

I think the changes improve correctness, but I want to make sure you're not
relying on the particulars I changed.

Adam-Vandervorst:
Do you have a plain text description of the change?

luketpeterson:
> Do you have a plain text description of the change?

The implementation without the change accumulates (accretes) child zippers in the `secondary` vec as the zipper moves, because factors that are ascended out of aren't removed during an ascend_until or reset.  Ordinary ascend was already fine.

Also the `factor_count` method is effectively meaningless so I got rid of it.

luketpeterson:
@Adam-Vandervorst Can you run compliance tests on this when you get a chance.  I think it fixes real bugs in the functionality but end-to-end tests in-situ with the client code should be done before merging.

Not urgent.  But I'm trying to get back on top of MORK maintenance after slacking off a bit during the past months.

Adam-Vandervorst:
@luketpeterson I'm traveling, have you checked `mork test` runs cleanly yet? I believe there only used in a prefixzipper so only the traits the nested zipper requires are exercised.

luketpeterson:
On my local machine, with this change, `cargo run -p mork -- test` finishes without errors (after setting `ACT_PATH` to  `"/tmp/"`)

Adam-Vandervorst:
Cool. You can merge it.

PR #57, opened 2026-07-28
Line list node graft fixes

imlvts:
There's a bug when doing `graft` on a `LineListNode`.
Grafting puts the child in slot 0; the value added afterwards lands in slot 1 under the same key.
That results in `LineListNode` iteration not getting the values/children properly.
This PR fixes the bug, and adds a check for `validate_node` to find similar issues.


luketpeterson:
Good catch.  That damn pair format is so complicated.

I've often felt that the "same key" state should have a prescribed order between the val and onward link, but I just tried both implementations and it appears to be more efficient to let it remain agnostic.

There was one unreachable case that I deleted, but otherwise good fixes.

PR #58, opened 2026-07-28
Add ArenaCompactTree::from_zipper_cached, re-using shared subtries

imlvts:
`from_zipper` walks every path of the source trie, so a subtrie that is
structurally shared (grafted in many places) is serialized once per
occurrence.  `from_zipper_cached` keys a map on the source's `shared_node_id`
and hands the already-built `Node` back instead of walking the subtrie again.

A cached `Node` can be handed to any parent because every `NodeId`/`LineId` in it
addresses arena data that is already written, and all arena references point
backwards.  The node itself still has to be copied into each sibling run --
the format requires siblings to be contiguous, so a shared subtrie cannot be
pointed at from two runs -- but the copy keeps pointing at the original line
data and children, so a repeated subtrie costs one node instead of a subtrie.

The traversal is the jumping catamorphism unrolled, rather than a call to
`into_cata_jumping_cached`, because that one only consults its cache one byte
below a fork.  Checking at the fork we land on after jumping a chain is what
catches a subtrie grafted under a multi-byte path, and re-using it needs an
operation the generic jumping algebra can't express ("prefix an already
folded subtrie"), but which is just a line node here.

For the parts of a trie with nothing to re-use both builders emit the same
bytes; tests assert that byte-for-byte over chains, wide branches, values at
every step, and random tries, and assert equal content plus a >50x size drop
on shared ones.

There's a test which stores 256^5 (or 10^12) paths in 37KB serialized.

### DIFF
--- a/src/arena_compact.rs
+++ b/src/arena_compact.rs
@@ -1367,6 +1367,10 @@ struct CachedFrame {
 /// - the cached cata's algebra is an `Fn`, so writing to the arena from it
 ///   would need interior mutability.
 ///
+/// TODO: GOAT: introduce an abstraction/modify `into_cata_jumping_cached`,
+///  to address the problems listed above.  The suggested API is to have
+///  `FnMut` variant for the caching catamorphism.
+///
 /// Sharing that begins in the middle of a jumped chain is still missed, since
 /// finding it would mean giving up `descend_until` and stepping byte by byte.
 ///
@@ -3365,6 +3369,10 @@ mod tests {
     /// Create a fully-populated map with depth `depth`.
     /// The nodes are maximally shared.
     /// There are 256**depth paths in the resulting map.
+    ///
+    /// TODO: Use `crate::utils::ints::gen_int_range` instead?
+    ///  Reason it was not used -- in the moment, the promise of sharing nodes
+    ///  in the `gen_int_range` was not clear.  This uses sharing explicitly.
     fn make_fully_populated_shared(depth: usize) -> PathMap<u64> {
         if depth == 0 {
             let mut map = PathMap::new();
--- a/src/arena_compact.rs
+++ b/src/arena_compact.rs
@@ -901,6 +901,71 @@ impl ArenaCompactTree<Vec<u8>> {
         build_arena_tree(zipper, map)
     }
 
+    /// Construct [ArenaCompactTree] from a read zipper, re-using the subtries
+    /// that are structurally shared in the source trie.
+    ///
+    /// This behaves exactly like [`Self::from_zipper`], except that whenever
+    /// the source zipper reports the same
+    /// [`shared_node_id`](ZipperConcrete::shared_node_id) for two positions
+    /// (i.e. both paths lead into the same in-memory node), the arena bytes
+    /// below that node are written only once, and the second occurrence points
+    /// back at them.  For tries built by grafting the same subtrie in many
+    /// places this can shrink the output (and the time to produce it) by
+    /// orders of magnitude; for a trie with no sharing it produces the same
+    /// bytes as [`Self::from_zipper`], at the cost of maintaining the reuse
+    /// map.
+    ///
+    /// ## Why the shared node is still copied
+    ///
+    /// The ACT format requires all children of a node to be laid out
+    /// contiguously, so that a parent only stores the offset of its *first*
+    /// child.  A shared subtrie therefore cannot be pointed at from two
+    /// different sibling runs: its top node has to live inside each run it
+    /// belongs to.  What gets re-used is everything the top node references —
+    /// its line data and its own children, which is where essentially all of
+    /// the bytes are.  So each additional occurrence of a shared subtrie costs
+    /// one copied node (a handful of bytes) rather than the whole subtrie.
+    ///
+    /// Note that a source position carrying a value is never reported as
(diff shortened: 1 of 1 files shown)


src/arena_compact.rs
```diff
@@ -2909,6 +3272,228 @@ mod tests {
         }
     }
 
+    /// Every (path, value) in `map`, collected with a catamorphism
+    fn map_contents(map: &PathMap<u64>) -> Vec<(Vec<u8>, u64)> {
```
adamv-symbolica: Even for testing, this seems like a silly helper since PathMap does implement iter.

src/arena_compact.rs
```diff
@@ -2909,6 +3272,228 @@ mod tests {
...
+    }
+
+    /// Every (path, value) in `act`, collected by walking it
+    fn act_contents(act: &ArenaCompactTree<Vec<u8>>) -> Vec<(Vec<u8>, u64)> {
+        let mut out = Vec::new();
+        if let Some(v) = act.get_val_at(b"") {
+            out.push((Vec::new(), v));
+        }
+        let mut z = act.read_zipper_u64();
+        while z.to_next_val() {
```
adamv-symbolica: Why is this different form the previous? Using into_cata_side_effect for PathMap and manual iteration for ACT.

src/arena_compact.rs
```diff
@@ -2909,6 +3272,228 @@ mod tests {
...
+                plain.get_data().len(), cached.get_data().len());
+        } else {
+            assert!(cached.get_data().len() <= plain.get_data().len(),
+                "{name}: cached={}B plain={}B", cached.get_data().len(), plain.get_data().len());
+        }
+    }
+
+    /// Graft `leaf` under every prefix, `levels` times over, so each level
+    /// shares the whole trie built by the level below it
+    fn make_shared_map(prefixes: &[&[u8]], levels: usize, leaf: PathMap<u64>) -> PathMap<u64> {
```
adamv-symbolica: This seems like an almost adversarially obscure function. 
Prefixes should be a map. Setting at all leaves should be product. 
Also consider using merkleize or int range instead.

src/arena_compact.rs
```diff
@@ -2909,6 +3272,228 @@ mod tests {
...
+        use crate::zipper::ZipperWriting;
+        let mut map = leaf;
+        for _level in 0..levels {
+            let mut next = PathMap::<u64>::new();
+            for prefix in prefixes {
+                let mut wz = next.write_zipper_at_path(prefix);
+                wz.graft(&map.read_zipper());
```
adamv-symbolica, line 3362: This should probably use graft_map

src/arena_compact.rs
```diff
@@ -1117,6 +1182,304 @@ fn build_arena_tree<V, Z, F>(zipper: Z, map_val: F) -> ArenaCompactTree<Vec<u8>>
...
+///   of an already-folded subtrie" — but which is trivial here: a line node
+///   pointing at the re-used node.
+/// - the cached cata's algebra is an `Fn`, so writing to the arena from it
+///   would need interior mutability.
+///
+/// Sharing that begins in the middle of a jumped chain is still missed, since
+/// finding it would mean giving up `descend_until` and stepping byte by byte.
+///
+/// See [`ArenaCompactTree::from_zipper_cached`] for the user-facing docs.
+fn build_arena_tree_cached<V, Z, F>(mut z: Z, map_val: F) -> ArenaCompactTree<Vec<u8>>
```
adamv-symbolica, line 1378: This should *really* give us a pause about the caching morphisms. Raison d'être (1) feels like it should be expressible in the caching catamorphism, though I understand it is not at the moment, (2) cached acata should just provide an unsafe FnMut: this function also needs an explanation why caching and mutability can be combined.

src/arena_compact.rs
imlvts: merkleize or int range won't work on the fully populated 5-byte example which has over 1 trillion paths.
I will try to find a better API.

Adam-Vandervorst: Why would int range not work? It does 2^128 paths instantly.

src/arena_compact.rs
imlvts: dropped. implemented `ArenaCompactTree::iter` instead.

src/arena_compact.rs
imlvts: same as above, moved to `ArenaCompactTree::iter`.

src/arena_compact.rs
imlvts: No. Using `graft_map` consumes the map, and it makes me think it will clone the Node -> no reuse of deepest maps. Moved two functions together for visibility.

src/arena_compact.rs
imlvts: This was one of the reasons I postponed the implementation of this feature -- I tried using cached morphisms, and failed to express it cleanly with the existing API. I will experiment on the APIs to see what I can do.

src/arena_compact.rs
imlvts: Checked it, didn't know it was working like that already.

adamv-symbolica:
Do we want to merge this as-is and leave the abstraction to later work?

imlvts:
I will mark two TODOs and merge as-is

PR #69, opened 2026-09-02
Lean model and differential fuzzer.

imlvts:
An executable specification of `PathMap`, `ReadZipper` and `WriteZipper` in
Lean 4 (`lean/`), and a harness that runs the same generated program against
the model and against the crate and diffs the traces.

## Layout

- `lean/PathMapModel/` — the model. Laws are proved or checked at `lake build`.
- `lean/differential.py` — the driver. `shrink.py` minimises a failing input;
  `mutate.py` checks the harness catches deliberate model corruption.
- `differential/` — a workspace crate with the Rust side: the decoder and
  trace renderer, the resident-process protocol, repro codegen, and the
  binaries `pathmap_trace`, `act_trace`, `zipper_bug_repros`,
  `act_bug_repros`, `act_merge_check`, `sharing_check`.
- `lean/FINDINGS.md` — the defects found, each with a standalone reproducer.

The `fuzz/` cargo-fuzz crate is deleted; `pathmap_trace --check` covers what it did.
Having two fuzzers, one of which does nothing is confusing.

## Run

```
# Install lean (once)
curl https://elan.lean-lang.org/elan-init.sh -sSf | sh && source ~/.elan/env
# Run the fuzzer
cd lean && lake build
cargo build --release -p differential
./lean/differential.py --random 500 --seed 1
./lean/differential.py --act --random 500 --seed 1
cargo run -p differential --bin zipper_bug_repros -- --list
```

## Notes

Spec has known warts. Primarily, because it tries to follow what PathMap does, to an extent.
We can discuss what it's supposed to do, or postpone that decision. The spec+fuzzer is useful in either case.

The fuzzer driver is written in Python, it doesn't matter much, but it can be rewritten in Rust.
The fuzzer skips certain ops because the library would fail so much the fuzzer output would be useless.
The ops will be enabled as the bugs are fixed, unlocking more coverage.


imlvts:
Known spec warts: [long version](https://gist.github.com/imlvts/948d13c53412c219ac7041abb6d858e7)

1. `AlgStatus.merge` it's copied from ring.rs, but not separately derived
3. `u64Ops` copies the crate's left-biased `u64` lattice, where `join` and `meet` ignore the second operand. The Element paths of the model's result resolution are never exercised by that instance.
4. `prunePath` is "logically inconsistent" and will be masked out
5. `toNextKPath` picked one of two disagreeing implementations. follows the native `ReadZipper` fallback to the zipper root when the focus is shallower than k, where the trait default returns false. 
6. `dropHead` discards values at depth exactly k. Nothing documents that, and the fixtures confirming it are copied from the crate's own tests.
7. `joinMapInto` reproduces two asymmetries with `join_into`: a short-circuit that throws away the value status after writing the value, and a "node exists" test where join_into tests "node empty". The harness masks its status most of the time anyway.
8. `meet2` never reports Identity because a comment in the implementation says so, even though the doc for the status says it could.

Adam-Vandervorst (approved):
LGTM! Thanks for the initiative!

PR #70, opened 2026-09-02
Fix join_into dropping the source when the destination node is empty

imlvts:
Joining a non-empty source into an *empty* destination map reported `Identity` and wrote nothing; the same join into a destination holding one unrelated key worked.  `try_as_tagged` answers `Some` for a `BorrowedRc` whatever it holds, so an empty destination node took the `pjoin_dyn` path and came back `Identity(SELF_IDENT)`: "the result is what you already have", which for an empty destination is nothing.

The function already draws the distinction between "no node" and "an empty node" for the *source*, with an explicit `node_is_empty` check in the branch above; it did not draw it for the destination.  Now it does, and an empty node takes the same path as a missing one, because the union of nothing with the source is the source.  The model agrees: `join_into` with an empty `self` node joins to the source and reports `Element`.

Regression test: `write_zipper_join_into_empty_destination`, into an empty map, into a dangling path, and into a non-empty map.  Fails before this change with the data missing.


luketpeterson:
Thanks!  Good fix.

PR #71, opened 2026-09-02
Fix a TrieRef slice underflow when its key crosses a node boundary

imlvts:
`TrieRef::new_with_key_and_path_in` descends one step before handing the rest of the walk to `node_along_path`, and assumed the child it found was reached at or beyond the end of `node_key`.  In debug that assumption was a `debug_assert!(consumed_byte_cnt >= node_key_len)`; in release it was `&path[consumed_byte_cnt - node_key_len..]`, an unchecked subtraction on `usize`.

The assumption does not hold.  `create_path` materialises an empty node part-way along a key, so `node_get_child` finds a child *inside* `node_key` rather than past it, and the release build computed `2 - 6` and indexed a slice at 18446744073709551612.  From the public API:

    map.create_path(&[2, 2]);
    let mut wz = map.write_zipper_at_path(&[2, 2]);
    wz.descend_to(&[2, 1, 3, 1]);
    wz.val_at(&[1, 1]);          // panic

When the child is reached inside `node_key`, what remains to walk is `node_key[consumed..] ++ path`, which is a slice of neither.  It does not need to be: `node_along_path` below already walks a whole key from a node and crosses boundaries itself, so that case now falls through to it, exactly as the "no child at all" case already did.  Both call sites had the same code and both are fixed.  This was the single largest panic site the differential fuzzer found, 37% of a sample.

Regression test: `trie_ref_key_crossing_a_node_boundary`, with nothing below the empty node and with values below it, through a write zipper and through `trie_ref_at_path`.  Fails before this change.

### DIFF
--- a/src/trie_ref.rs
+++ b/src/trie_ref.rs
@@ -165,12 +165,28 @@ impl<'a, V: Clone + Send + Sync + 'a, A: Allocator + 'a> TrieRefBorrowed<'a, V, 
                 core::slice::from_raw_parts(temp_key_buf.as_mut_ptr().cast::<u8>(), total_buf_len)
             };
 
-            if let Some((consumed_byte_cnt, next_node)) = node.as_tagged().node_get_child(next_node_path) {
-                debug_assert!(consumed_byte_cnt >= node_key_len);
-                node = next_node;
-                path = &path[consumed_byte_cnt-node_key_len..];
-            } else {
-                path = next_node_path;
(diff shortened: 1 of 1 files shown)


luketpeterson:
It seems TrieRef was carrying some cruft from an earlier design.  I'm doing a quick pass to simplify the implementation.

luketpeterson:
There were a number of edge cases in TrieRef that this PR cleans up.  Merging.

PR #72, opened 2026-09-02
Make ACTZipper::reset return to an off-trie root instead of a real ancestor

imlvts:
A zipper can be *rooted* at a path that does not exist: `fork_read_zipper` at an off-trie focus makes one, and so does `read_zipper_at_path` with such a path.  `reset` put the path back but then cleared `invalid` unconditionally, so the zipper came back believing it was rooted on the trie, at whatever real node the stack still pointed at, its deepest existing ancestor.  It then answered `val()` with that ancestor's value:

    trie: root=38, [1,0,2]=22, [1,1]=72
    fork at [1,1,3] (does not exist): val None  ->  after reset(): Some(72)
    fork at [9]     (does not exist): val None  ->  after reset(): Some(38)

The zipper knew where its root was (`origin_depth`, `origin_node_depth`) but not how far off the trie it was, so `reset` had nothing to restore. `origin_invalid` records that, `with_root_here` captures it when a fork takes the focus as its root, and `reset` restores it.

Regression test: `act_zipper_reset_returns_to_an_off_trie_root`, forks at three off-trie paths and a zipper created at one.  Fails before this change.

### DIFF
--- a/src/arena_compact.rs
+++ b/src/arena_compact.rs
@@ -2307,9 +2307,11 @@ where Storage: AsRef<[u8]>
     cur_node: Node,
     stack: Vec<StackFrame>,
     path: Vec<u8>,
+    pub(crate) invalid: usize,
+    // origin_* fields are used to remember the position of zipper's root
     origin_depth: usize,
     origin_node_depth: usize,
-    pub invalid: usize,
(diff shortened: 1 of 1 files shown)


luketpeterson:

PR #75, opened 2026-09-02
Fix meet_2 unwrapping a None when a source focus holds an empty node

imlvts:
`meet_2` guards both sources with `try_as_tagged`, then on the `Identity` arm unwraps `into_option` on the same value.  The two disagree: `try_as_tagged` answers `Some` for a `BorrowedRc` whatever it holds, while `into_option` answers `None` when that node is empty.  So a source whose focus holds an empty node passes the guard and panics on the unwrap.  An empty node is what `create_path` leaves behind, so a read zipper rooted at a dangling path is enough:

    a.create_path(&[0]);
    wz.meet_2(&a.read_zipper_at_path(&[0]), &b.read_zipper_at_path(&[0]));

The mask on the `Identity` arm says which *operand* the intersection equals, so an empty operand means an empty intersection: graft nothing and report `None`.  That is what the model specifies for `meet_2` when either source node is empty.

Regression test: `write_zipper_meet_2_with_an_empty_source_node`, three shapes for the other operand, both operand orders.  Fails before this change with an `Option::unwrap()` panic.


luketpeterson:
I updated comments but the code change was correct and the test was good.

PR #80, opened 2026-09-02
NO MERGE ACTZipper::val_count test, to be included in cata rewrite

imlvts:
`val_count` opened with `zipper.reset()`.  `ZipperValues::val_count` counts the values at and below the *focus*, so this returned the count for the whole subtrie under the zipper's root wherever the focus was, and was right only when the focus happened to be at the root.

Removing the reset is not quite enough: `to_next_val` walks the zipper's entire subtrie rather than the part below the focus, so the walk has to stop when it leaves.  Depth-first order visits everything below the focus before anything outside it, so the first path that no longer starts with the focus ends the count.

Regression test: `act_zipper_val_count_counts_from_the_focus`, against the PathMap zipper at six foci from the map root and three from a zipper rooted below it.  Fails before this change.

lean/FINDINGS.md ACT finding "val_count_ignores_focus" on the lean-fuzzer-restage branch.

### DIFF
--- a/src/arena_compact.rs
+++ b/src/arena_compact.rs
@@ -2920,13 +2920,24 @@ where Storage: AsRef<[u8]>
     /// WARNING: This is not a cheap method. It may have an order-N cost
     fn val_count(&self) -> usize {
         timed_span!(ValueCount, COUNTERS);
+        //`ZipperMoving::val_count` counts the values at and below the *focus*.  This used to
+        // `reset()` first, so it counted the whole subtrie below the zipper's root and returned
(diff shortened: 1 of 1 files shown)


luketpeterson:
The current val_count pathway is totally gone, once we merge the new catamorphism trait.  So I'd rather not merge the behavior change because it'll just create a conflict.

The test might be fine under the "low bar for a test" philosophy.  Also I was thinking I ought to add val_count benchmarks for ACT.

So I'll keep this PR open for the test and the yet-to-be-added benchmarks

imlvts:
duplicate of #80

Issue #85, opened 2026-09-03
Four more dangling path issues


adamv-symbolica:
Common helpers for the snippets:

```rust
use pathmap::PathMap;
use pathmap::zipper::*;

fn mk(keys: &[&[u8]]) -> PathMap<()> {
    let mut m = PathMap::new();
    for k in keys { m.set_val_at(k, ()); }
    m
}
fn keys(m: &PathMap<()>) -> Vec<String> {
    m.iter().map(|(k, _)| String::from_utf8_lossy(&k).into_owned()).collect()
}
/// {ca, cb, d} with "c" turned into a dangling branch
fn dangling_c() -> PathMap<()> {
    let mut m = mk(&[b"ca", b"cb", b"d"]);
    { let mut wz = m.write_zipper(); wz.descend_to(b"c"); wz.remove_branches(false); }
    assert_eq!(keys(&m), ["d"]);
    m
}
```

---

## Mutating below a dangling branch of a DenseByteNode asserts "Attempted to make_unique on an empty sentinel node"

**Summary.** Any write-zipper operation whose path continues *below* a dangling byte of a dense node panics in
`make_unique`. The same operation on a pair-node (LineListNode) parent works, and mutating exactly *at* the
dangling byte works, so this is specific to the dense node's child lookup.

**Reproduction.**

```rust
#[test]
fn set_val_below_dangling_byte_of_dense_node() {
    let mut m = dangling_c();                 // {d} plus dangling "c", root is a pair node
    m.set_val_at(b"e", ());
    m.set_val_at(b"f", ());                   // root now has c, d, e, f: upgraded to a dense node
    m.set_val_at(b"cx", ());                  // panics
    assert_eq!(keys(&m), ["cx", "d", "e", "f"]);
}
```

Same panic with a graft below the dangling byte:

```rust
let mut wz = m.write_zipper(); wz.descend_to(b"cxy"); wz.graft_map(mk(&[b"z"]));
```

Controls that pass: `dangling_c()` then `set_val_at(b"cx")` directly (pair-node parent), and
`set_val_at(b"c")` after the dense upgrade (mutation at the dangling byte itself).

**Expected.** `{cx, d, e, f}`.

**Actual.**

```
panicked at src/trie_node.rs:3063:13:
Attempted to make_unique on an empty sentinel node
```

**Root cause.** `ByteNode::node_get_child_mut` (dense_byte_node.rs:690) returns the child rec for any byte in
[...]

PR #86, opened 2026-09-03
Fix panics on dangling paths: issue #85 and the empty-sentinel joins

imlvts:
(addressing #85 + blast radius)

A dangling path (a location that exists in the trie but has no value and nothing
below it, as left by `remove_branches(prune = false)`, `remove_val(prune = false)`,
or `create_path`) is stored as a child link to the shared empty sentinel node. The
sentinel has no refcount word, so it cannot be made mutable, and several code paths
reached `make_mut` on it, unwrapped a node that is not there, or hit an
`unreachable!()` that assumed both operands of a join were non-empty. This branch
makes every operation that can meet a sentinel treat it as "nothing here" and
replace it rather than mutate it. It covers all four items of issue #85 plus the
join and drop-head paths that the same representation breaks.

## Joins and drop_head over a dangling path (commit 1)

- `TrieNodeODRc::join_into`: joining an empty node into anything is the identity;
  joining anything into an empty node replaces it. `DenseByteNode::join_into_dyn`
  and the CoFree join route through it, so a dangling child on either side no
  longer reaches `make_mut`.
- `ByteNode::pjoin` / `join_into`: two dangling paths at the same byte join to a
  dangling path instead of `unreachable!()`.
- `drop_head_dyn` (dense and list): a dangling child contributes nothing to the
  dropped-head result and is skipped instead of made mutable.
- `merge_list_nodes`: when both nodes are empty it returns `None` before touching
  `entries[0]`; reading it was an `assume_init` on uninitialised memory, and
  dropping the garbage payload corrupted an arbitrary refcount (the test binary
  aborted with SIGABRT).
- `LineListNode::pjoin_dyn` against a dense node: two empty nodes join to an empty
  result instead of `unreachable!()`.
- `LineListNode::factor_prefix`: a value at a multi-byte key sharing its first byte
  with the other slot is no longer called legal; `drop_head` could leave
  `(Val@"aa", Val@"ab")`, a shape insertion never creates and indexed descent
  cannot traverse, and it is now factored into `Child@"a" -> {a, b}`. `drop_head`
  also merges two keys that coincide after shortening, which used to leave one
  path with two values (iteration saw it once, `val_count` twice).
- `join_k_path_into`: an empty result clears the branch instead of grafting an
  empty node.
- Tests: dangling children under pair-node and dense parents, joins of emptied
  roots, `join_k_path_into` over dangling children, and a randomised algebra
  program that checks node invariants after every step.

## graft_masked_branches at a dangling focus (commits 2 and 3; issue #85 item 3)

- With three or more mask bits the method split the focus before looking at the
  source, and at a dangling tip or empty root the split has no node to hand back:
  `try_borrow_focus_mut().unwrap()` on `None`. Masked branches the source lacks
  are now removed up front in every arm, as the doc comment specifies, only
  branches the source has are grafted, and a focus that is an empty stub gets a
  fresh node to merge into.
- The one- and two-bit arms previously grafted whatever the source had at each
  masked byte, which for an absent branch left a dangling child behind, while the
  three-bit arm removed it; the arms now agree.

## Write zipper reads after the root is emptied (commit 4)

- `MutNodeStack::top()` returned `None` for an empty root while `top_mut()`
  returned the node, so after `remove_branches` at the root (or a
  `join_k_path_into` that collapsed everything) the next `path_exists` unwrapped
  `None`. `top()` now returns the empty node, which every reader already handles.

## Issue #85 items 1, 2, 4 (commits 5 to 7)

- `ByteNode::get_child_mut` reports an empty rec as "no child", as
  `LineListNode::get_child_mut` already does. A pair node upgraded to a dense node
  carries its dangling slot over as such a rec, and the descent called `make_mut`
  on it for any operation continuing below the dangling byte
  (`set_val_at(b"cx")`, a graft at `cxy`, `restrict`, `remove_unmasked_branches`).
  The descent now stops at the parent and the following mutator replaces the rec
  through `set_child`.
- `join_into_take`: an empty taken source leaves the destination alone and answers
  `Identity`; an empty taken destination is replaced by the source. Both
  previously handed the sentinel to `make_mut` or to `graft_internal`.
- `LineListNode::restrict_slot_contents`: a child-link slot whose key leads to a
  dangling byte of the other operand is dropped, mirroring
  `subtract_from_slot_contents`; calling `as_tagged` on the missing node was an
  explicit panic in `PathMap::restrict` and both zipper forms.
- Each carries the reproductions from the issue, including its controls, as
  regression tests.

## Verification

- `cargo test --release`: 848 tests pass.
- Differential fuzzing against the Lean model, same generator and seeds as the
  master baseline: crate run 14786 -> 15105 of 20000 programs agreeing; ACT run
  2357 -> 2407 of 5000. The "make_unique on an empty sentinel" class (344 programs
  on master) no longer occurs. The three divergences new to this branch are
  programs that previously died earlier on a panic fixed here.

## Not in this branch

- The dense `drop_head_dyn` still folds children from the highest byte, so a value
  collision in `join_k_path_into` keeps the highest k-path's value where the list
  node and the specification keep the lowest. Unrelated to the sentinel and left
  as is.

### DIFF
--- a/src/dense_byte_node.rs
+++ b/src/dense_byte_node.rs
@@ -295,11 +295,6 @@ impl<V: Clone + Send + Sync, A: Allocator, Cf: CoFree<V=V, A=A>> ByteNode<Cf, A>
 
     #[inline]
     fn get_child_mut(&mut self, k: u8) -> Option<&mut TrieNodeODRc<V, A>> {
-        //An empty rec is a dangling byte: the byte is in the mask but there is nothing to descend
-        // into, and the empty sentinel it holds cannot be made mutable.  It is reported as "no
-        // child", as LineListNode::get_child_mut does, so a descent stops at this node and the
-        // mutator that follows (node_set_val, node_set_branch, a graft) replaces the rec through
-        // set_child instead of calling make_mut on it.
         self.get_mut(k).and_then(|cf| cf.rec_mut()).filter(|child| !child.is_empty())
     }
 
@@ -1312,8 +1307,6 @@ impl<V: Clone + Send + Sync, A: Allocator, Cf: CoFree<V=V, A=A>> TrieNode<V, A> 
     }
 
     fn join_into_dyn(&mut self, mut other: TrieNodeODRc<V, A>) -> (AlgebraicStatus, Result<(), TrieNodeODRc<V, A>>) where V: Lattice {
-        //An empty `other` (e.g. the sentinel behind a dangling path) contributes nothing, and it
-        // cannot be made mutable; this mirrors the `EMPTY_NODE_TAG` arm in LineListNode::join_into_dyn
         if other.as_tagged().node_is_empty() {
             return (AlgebraicStatus::Identity, Ok(()))
         }
@@ -1353,7 +1346,6 @@ impl<V: Clone + Send + Sync, A: Allocator, Cf: CoFree<V=V, A=A>> TrieNode<V, A> 
                 //WARNING: Don't be tempted to swap the node itself with its first child.  This feels like it
                 // might be an optimization, but it would be a memory leak because the other node will now
                 // hold an Rc to itself.
-                //A dangling child (the empty sentinel) has nothing below the dropped byte and can't be made mutable
                 match self.values.pop().unwrap().into_rec().filter(|child| !child.is_empty()) {
                     Some(mut child) => {
                         if byte_cnt > 1 {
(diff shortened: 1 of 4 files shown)


luketpeterson:
There was a lot in here.  Most of it good, a few things (like the change to MutCursor::top) not good.

But the biggest issue seems to be that the performance of graft_masked_branches regressed 2x or worse.  So trying to get to the bottom of that.  Once we get that speed back then this PR is good to merge.

luketpeterson:
Man... The clanker wrote a whole paragraph about how the original code didn't follow the spec.  I was a little skeptical because I wrote that code myself only 6 weeks ago, I spent a lot of time thinking about it, and it was fresh in my mind.

But the clanker made a convincing argument and I bought it at first.  But its "fix" was to effectively do all the work twice.  The benchmark regression had me look a lot closer at the implementation and realize a huge chunk of the changes it made were totally unnecessary!

Issue #90, opened 2026-09-06
Merkleize byte-identical node allocation


adamv-symbolica:
```rs
use pathmap::PathMap;
use pathmap::viz::{viz_maps, DrawConfig, VizMode};

fn main() {
    // Every bitstring over {0,1} of length 1..=4 ending in 1, plus the empty path (16 paths)
    let mut paths: Vec<Vec<u8>> = vec![vec![]];
    for len in 1..=4usize {
        for bits in 0..(1u32 << len) {
            let p: Vec<u8> = (0..len).map(|i| ((bits >> (len - 1 - i)) & 1) as u8).collect();
            if *p.last().unwrap() == 1 { paths.push(p); }
        }
    }
    let mut map = PathMap::from_iter(paths.iter().map(|p| (p.as_slice(), ())));

    let count = |m: &PathMap<()>| {
        let dc = DrawConfig { mode: VizMode::Mermaid, ascii_path: false, hide_value_paths: true,
                              minimize_values: true, logical: false, color: false };
        let mut out = Vec::new();
        viz_maps(std::slice::from_ref(m), &dc, &mut out).unwrap();
        String::from_utf8(out).unwrap().lines().filter(|l| l.contains("shape: rect")).count()
    };

    println!("before: {}", count(&map));   // 15
    let r = map.merkleize();
    println!("after: {} ({r:?})", count(&map)); // 7, expected 4
}
```

PR #91, opened 2026-09-06
Fix values being incorrectly attributed to nodes in merkleize hash

imlvts:
This is supposed to address #90


imlvts:
Closing in favor of another change

PR #102, opened 2026-09-11
fixed the code to not set a bit that was already cleared.

ClarkeRemy:
The addition of the helper functions is to allow for implementing more const constructed masks in the future.

### DIFF
--- a/src/utils/mod.rs
+++ b/src/utils/mod.rs
@@ -577,6 +577,15 @@ pub trait BitMask {
     fn not(&self) -> Self where Self: Sized;
 }
 
+#[inline] const fn u64_4_raw_set_bit(mask : &mut [u64;4], k : u8) {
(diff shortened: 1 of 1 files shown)


src/utils/mod.rs
```diff
@@ -577,6 +577,15 @@ pub trait BitMask {
     fn not(&self) -> Self where Self: Sized;
 }
 
+#[inline] const fn u64_4_raw_set_bit(mask : &mut [u64;4], k : u8) {
```
luketpeterson: Why are we operating on the insides of ByteMask, as opposed to on ByteMask?

luketpeterson: Are you thinking these new const functions will be pub, or is this just a part of a larger change?

luketpeterson:
Good fixes.  I  added a bunch more testing and caught another bug.  Merging.

PR #103, opened 2026-09-14
Added const inclusive bit range primitives.

ClarkeRemy:
These are const methods for defining and modifying ByteMasks using ranges.
They should be efficient enough that one could use them with runtime values.
There are three range methods, set, clear, and compliment.


luketpeterson (approved):
I made some superficial (but API-changing) tweaks to harmonize it to the vocabulary and conventions of the rest of the API.

Good to merge, unless you want to discuss further.

PR #106, opened 2026-09-16
Fix val_at returning the focus value for a dangling child

imlvts:
TrieRef lookups stepped into a child node even when that consumed the whole path, and an empty remaining path answers with the focus value. Only step into the child when path remains.


imlvts:
https://github.com/Adam-Vandervorst/PathMap/actions/runs/35157988323#summary-105006282703

Benchmark: +27.3% binary_zipper_iter/50. doesn't seem to be noise, but doesn't affect overall benchmark results?

PR #107, opened 2026-09-16
Keep a value in subtract when the other side only passes through

imlvts:
ByteNode::psubtract_abstract dropped a value at a byte where the other node has no value, only a longer path. The value now survives.


imlvts:
https://github.com/Adam-Vandervorst/PathMap/actions/runs/35163010909#summary-105018138331

Benchmark: +57.2% superdense_drop_bench/200, but overall results are flat?

luketpeterson:
Good fixes.  Merging

PR #109, opened 2026-09-16
Fix write zipper node stack after prune_path

imlvts:
prune_path walked the node stack up to find where pruning stops but left it there when the zipper did not move, so the next write through the focus went to the wrong node (get_val_or_set_mut panicked). Walk the stack back down.


imlvts:
https://github.com/Adam-Vandervorst/PathMap/actions/runs/35158040227#summary-105053523842

Benchmark: +29.6% borrowed_head_write_creation_cleanup/10, and zipper_head_owned+10.5%

luketpeterson:
The clanker used non-standard terminology to describe what happened - The code describes this state as the zipper as being deregularized.  But it's the right fix.

Merging.

PR #112, opened 2026-09-17
Fix remove_unmasked_branches at a dangling path

imlvts:
A dangling path is a link to the empty node, which `get_child_mut` won't return, so the zipper passed the stub's own key to the parent node as a branch and tripped a LineListNode debug assert. There is nothing below a dangling path to filter, so do nothing.


luketpeterson:
Simplified the fix a lot.  I honestly don't know what the clanker was thinking with this fix.  Like a 20 line preflight block instead of adding one extra check (that was part of the 20 line block).

Anyway, simplified the fix.  Merging.

PR #113, opened 2026-09-17
Fix ACTZipper sibling steps and ascend_until off the trie

imlvts:
to_sibling answered None for a focus one byte off the trie, so to_next_step gave up instead of moving to the next sibling. ascend_until then checked the landing node rather than the focus for a value or branch, stopping at the wrong ancestor.


imlvts:
To be changed: #111 includes the fix for sibling movement, separate sibling step

PR #115, opened 2026-09-17
Return Identity from integer psubtract of unequal values

imlvts:
u64 and u16 psubtract returned Element(*self) when nothing was subtracted. Nodes propagate identity masks, so subtract_into reported Element for an unchanged trie. The Lean model's u64Ops follows.


luketpeterson:
@imlvts What is there a reason behind this change?  If there is an expedient reason to change the behavior then it's fine.

My issue is that the policies implemented in the Lattice trait implementations on primitives are basically nonsense.  Or at least unjustified.

The plan is to clean them up with well-behaved policies implemented via ZST policy objects.  I have a plan for how to go about that.  But any kind of near-term tweak to the policy gives the impression of a well reasoned design, and creates inertia to the current behavior.  Encoding anything about the current Lattice impls in the lean model disturbs me greatly.

But again, if you have a pragmatic reason then we can merge this PR, as-is.

PR #116, opened 2026-09-17
Fix join_into overwriting a destination that holds the source

imlvts:
Joining with a TinyRefNode swapped operands without inverting the identity mask, so an unchanged destination reported COUNTER_IDENT and join_into replaced it with the source. merge_guts and the integer pjoin also under-reported identities, giving Element for unchanged joins.


luketpeterson:
3 fixes in here.  2 of them are very good (carrying status through merge_guts, and the dispatch to tiny_node).  The third is harmless (changing policy on trash Lattice impls).  Merging because I don't care enough to break out the third fix into a separate PR if it's going to get merged anyway.  And I figured there must be a reason behind the change and don't really care enough to fight about it.

PR #121, opened 2026-09-19
descend_first_k_path(0) compliance fixes

luketpeterson:
descend_first_k_path(0) behavior clarification exposed a bunch of out-of-spec implementations.

Adding test to ZipperIteration test macro to test compliance with clarified spec on `descend_first_k_path(0)` behavior, and fixing all implementations that were out of spec


imlvts:
This might allow us to unmask some things in fuzzer and spec

imlvts:
Unmasking k=0 has revealed a few discrepancies:

https://github.com/Adam-Vandervorst/PathMap/compare/bugfix/descend_k_path_zero_spec...imlvts:PathMap:bugfix/descend_k_path_zero_spec

Note: the fuzzer might regress because we're unmasking new cases, and it's fine.

Adam-Vandervorst:
I'm struggling to see what this would mean -- why not just disallow 0 as an argument?

PR #126, opened 2026-09-23
Fix ZipperHead exclusive path at the head's own root

imlvts:
With an empty path and the head's root inside a node, the walk popped a byte of the head's own path and never put it back. The next request then read past its key buffer (UB in release) or panicked.


luketpeterson:
I see.  That ZH code was assuming it had a rooted WriteZipper.  oops.  Good fix.

PR #127, opened 2026-09-23
Fix zipper_head on a write zipper with a borrowed path

imlvts:
as_static_path_zipper asserted the zipper held no borrowed origin path, but write_zipper_at_path always does. Once buffers are prepared the path lives in prefix_buf, so drop the borrowed slice instead.


luketpeterson:
Good catch.  I ended up moving the fix to a slightly different spot.  Merging.

PR #128, opened 2026-09-23
Fix get_trie_ref on a read zipper that owns its root

imlvts:
ZipperHead read zippers own their root node, and get_trie_ref borrowed the root for the zipper's whole lifetime, which panics for an owned root. The TrieRef only needs to borrow the zipper.


luketpeterson:
Good fix.  Merging.

PR #129, opened 2026-09-23
Fix ProductZipper is_shared at a factor root

imlvts:
A secondary factor's root isn't a child of the node above it, so the core zipper's parent lookup unwrapped None. Ask the factor's TrieRef instead.

Only report sharing in the last factor, for both ProductZipper and ProductZipperG. In an earlier factor the subtrie below a node continues into the following factors, so the same node reached in different factors is a different subtrie of the product. A cached cata keyed by shared_node_id conflated the two and returned wrong results.


imlvts:
superseeded by #135

PR #135, opened 2026-09-23
Fix: ProductZipper crashes

imlvts:
Fix the default k-path walk looping forever at a leaf
With nothing below the base and no sibling, k_path_default_internal never
reached its exit and spun. It also let k = 0 step sideways from the base.
Stop when back at the base.

Fix ProductZipper factor bookkeeping at the root
A factor was entered below a primary path that doesn't exist, and a
sibling step at depth 0 dropped the factor record although the core
zipper can't leave the factor there. Later moves then saw factor roots
the zipper didn't know about, and is_shared unwrapped None.


src/product_zipper.rs
```diff
@@ -136,7 +136,7 @@ impl<'factor_z, 'trie, V: Clone + Send + Sync + Unpin, A: Allocator> ProductZipp
     /// `product_zipper_test4` for more discussion.
     #[inline]
     fn ensure_descend_next_factor(&mut self) {
-        if self.factor_paths.len() < self.secondaries.len() && self.z.child_count() == 0 {
+        if self.factor_paths.len() < self.secondaries.len() && self.z.child_count() == 0 && self.z.path_exists() {
```
adamv-symbolica: Hmm, this used to be there, but it cost performance and it wasn't clear why this was needed.

PR #136, opened 2026-09-23
Fix/product zipper shared node id by factor

imlvts:
There's an issue with sharing ProductZipper: cannot pass node id, since the same node can be used at different levels.
This is fixed by xoring factor level into the second byte of node id. With a sanity check.
The same approach does not work for DependentProductZipper[G], since a shared node can have different next factor depending on path. Added a test in case someone wants to do it.


src/product_zipper.rs
```diff
@@ -361,9 +375,55 @@ impl<'trie, V: Clone + Send + Sync + Unpin + 'trie, A: Allocator + 'trie> Zipper
...
+        return Some(id);
+    }
+    let tag = u8::try_from(factor + 1).ok()?;
+    let [.., below_top, top] = id.to_le_bytes();
+    // The assumption is that the address space is 48 bit, and the node id was not already tagged
+    // Skip sharing if that's not the case.
+    if below_top != top {
+        return None;
+    }
+    Some(id ^ ((tag as u64) << 48))
```
adamv-symbolica: We already use a crate for this which steals bits from alignment, sign, and address space?

imlvts: no, we don't use a crate. there's code copied from `ointers` in `src/trie_node.rs:2804`.

adamv-symbolica: >  In a canonical address the byte below the top byte equals the top

No? For physical addresses you can use the top byte, but for virtual ones you can't do that, and definitely not the top two bytes.

> below_top != top

`0b00000001 0b00000001` is a perfectly fine start of an untagged pointer.

> u8::try_from(factor + 1).ok()?;

Failing sharing on factor count seems devious.

imlvts: > Failing sharing on factor count seems devious.

of all the limitations that PathMap and MORK have, this one should be the least controversial.
if you have more than 254 factors in a product zipper, please let me know.
doing any sort of traversal there would take too much time.

adamv-symbolica: > there's code copied from ointers in ...

Hmm, why again wasn't this lifted out @luketpeterson ? Seems like we don't want to be doing this ad-hoc, a mistake like this could corrupt the trie.

> doing any sort of traversal there would take too much time.

We do traversals over 1000+ bytes all the time

> of all the limitations that PathMap and MORK have, this one should be the least controversial.

This should come with a big red exclamation mark. With that, we can move the <=255 assumption to optimize other methods as well perhaps.

imlvts: > No? For physical addresses you can use the top byte, but for virtual ones you can't do that, and definitely not the top two bytes.

I double-checked myself, and it's system-dependent. for x86, 4K tables, top 2 bytes of virtual memory are either 0x00 or 0xff.
On all systems, top 4 bits are sign-extended. I'll find a better way.

imlvts: > We do traversals over 1000+ bytes all the time

the limit is on the number of nested maps, not byte depth. so `[map1, map2, map3, ... map254]`

adamv-symbolica: > I double-checked myself, and it's system-dependent. for x86, 4K tables, top 2 bytes of virtual memory are either 0x00 or 0xff.

Reproduced with custom Alloc on x86, 4k tables, with 5-level paging that 0x0000_0020_0000_0000 and 0x0001_0020_0000_0000 are sent to the same node id, causing trie corruption. 

> the limit is on the number of nested maps

Unless enrolling factors is slow, search algorithms controlling zippers can definitely still use this, see utils int.

imlvts: Converting this to draft until we have a better solution.
If shared node id is u64 and they're essentially pointers, there will be still limits on what we can do with this.

